PR #71, opened 2020-08-31
minor code cleanups

dfa1:
Hello,

 thank you for this nice library. I would like to propose few minor fixes:

- missing Thread.currentThread().interrupt(); after catching InterruptedException
- converting Object.nonNull to Assertions.nonNull
- removing some vestigial @SuppressWarnings("unchecked")

### DIFF
--- a/src/main/java/org/dataloader/BatchLoaderEnvironment.java
+++ b/src/main/java/org/dataloader/BatchLoaderEnvironment.java
@@ -1,12 +1,12 @@
 package org.dataloader;
 
+import org.dataloader.impl.Assertions;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-
-import static java.util.Objects.nonNull;
 
 /**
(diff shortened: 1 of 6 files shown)


src/main/java/org/dataloader/BatchLoaderEnvironment.java
```diff
@@ -78,8 +78,8 @@ public Builder context(Object context) {
         }
 
         public <K> Builder keyContexts(List<K> keys, List<Object> keyContexts) {
-            nonNull(keys);
```
dfa1, line 81: here I believe the intent was to throw NPE, correct?

src/main/java/org/dataloader/impl/CompletableFutureKit.java
```diff
@@ -20,14 +20,15 @@
         return future;
     }
 
-    public static Throwable cause(CompletableFuture completableFuture) {
+    public static Throwable cause(CompletableFuture<?> completableFuture) {
```
dfa1, line 23: not sure about this raw-type

bbakerman (approved):
Thanks for this

PR #77, opened 2021-01-31
Add support for minimum batch size

sachinpkale:
- This change introduces two new options in DataLoaderOptions: `minBatchSize` and `maxWaitTimeInMillis`. 
- These options helps to configure minimum batch size to be formed before it is passed to BatchLoader.
- This is particularly helpful in scenarios where applications do not have control on when to call dispatch() method (for example, GraphQL).
- Both of these parameters need to be configured in order to achieve minimum batch. Configuring only one parameter will have the same impact of not configuring at all.


bbakerman (commented):
See #87 to do what you have but outside the DataLoader itself

PR #107, opened 2021-12-21
Upgrade to Gradle 7.3.2

berngp:
This commit upgrades the build to use Gradle 7.3.2, instead of the older Gradle 6.6.1.
As part if this upgrade we need to change the scope of some dependencies such that..

* `compile`, is now `api`.
* `testCompile`, is now `testImplementation`

Note that instead of using `api` you can use `implementation`.
More information available at [Gradle Docs].

[Gradle Docs]: https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separatio://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation


bbakerman (approved):
Thanks for this

PR #121, opened 2022-07-15
Copy across valueCache in DataLoaderOptions copy constructor

AlexandreCarlton:
This looks like an accidental omission, so we include this along with
all the other variables inside this class.


bbakerman (approved):
Good catch

PR #126, opened 2023-03-07
Prepend 0.0.0 to build version

dondonz:
Adding this so build versions do not appear as the "latest" release on Maven, when sorting in descending order


kaqqao:
Why not publish snapshots and not pollute Maven Central with random builds in the first place?

andimarek:
@kaqqao We publish every build to maven central because often users can't access certain builds without them being available in maven central.

Also: while not a release itself our builds are often used in production even before an official release, this is why they are not SNAPSHOTs.

PR #135, opened 2023-11-10
Lazily initialize Executor in ScheduledDataLoaderRegistry builder

### LINKED ISSUE #134: ScheduledDataLoaderRegistry.newScheduledRegistry() creates new ScheduledExecutorService on every invocation
Calling ScheduledDataLoaderRegistry.newScheduledRegistry() creates a new ScheduledExecutorService on every call, regardless of whether one is passed to the builder or not.

To reproduce:

```java
ScheduledExecutorService myScheduledExecutor = Executors.newSingleThreadScheduledExecutor();
ScheduledDataLoaderRegistry registry = ScheduledDataLoaderRegistry.newScheduledRegistry().scheduledExecutorService(myScheduledExecutor).build();
```

This is due to the builder field [being initialized with a default](https://github.com/graphql-java/java-dataloader/blob/5f8ec4abca7f54fd0371b728deff7594fac21d26/src/main/java/org/dataloader/registries/ScheduledDataLoaderRegistry.java#L260); it might be better to initialize the field to null and then only create a new ScheduledExecutorService if a custom one is not passed in.

kilink:
Calling ScheduledDataLoaderRegistry.newScheduledRegistry would create a new ScheduledExecutorService on every call, regardless of whether a custom one was supplied. Move creation of the default ScheduledExecutorService from the builder field to the build method to avoid the issue. Fixes #134

### DIFF
--- a/src/main/java/org/dataloader/registries/ScheduledDataLoaderRegistry.java
+++ b/src/main/java/org/dataloader/registries/ScheduledDataLoaderRegistry.java
@@ -257,7 +257,7 @@ public class ScheduledDataLoaderRegistry extends DataLoaderRegistry implements A
         private final Map<String, DataLoader<?, ?>> dataLoaders = new LinkedHashMap<>();
         private final Map<DataLoader<?, ?>, DispatchPredicate> dataLoaderPredicates = new LinkedHashMap<>();
         private DispatchPredicate dispatchPredicate = DispatchPredicate.DISPATCH_ALWAYS;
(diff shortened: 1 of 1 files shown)


srinivasankavitha:
This fix is important for services that have high RPS since the current implementation creates a new instance of the ScheduledExecutorService every time a request is processed.

bbakerman (approved):
Thanks for this fix - I agree this is better situation

bbakerman:
3.2.2 is now released

srinivasankavitha:
Will there also be a graphql-java release with this update?

bbakerman:
No

PR #136, opened 2023-11-10
Avoid allocations in DataLoaderHelper.dispatch when there's no work

### LINKED ISSUE #132: Avoid allocations when dataloader has no work
Our GraphQL server has many dataloaders registered. I noticed that each dataloader is dispatched, and lists are allocated even when the dataloader has no work to do.

https://github.com/graphql-java/java-dataloader/blob/master/src/main/java/org/dataloader/DataLoaderHelper.java#L163-L169

It would be great to avoid allocating these lists when the dataloader doesn't have work to do, especially when GraphQL queries are large (many calls to `dispatchAll()`, and there are many dataloaders registered that are sparsely used.

kilink:
Bail out early in DataLoaderHelper.dispatch when loaderQueue is empty to avoid unnecessary allocations; additionally, when there is work, size the allocated Lists precisely based on the loaderQueue size. Fixes #132

### DIFF
--- a/src/main/java/org/dataloader/DataLoaderHelper.java
+++ b/src/main/java/org/dataloader/DataLoaderHelper.java
@@ -162,12 +162,21 @@ class DataLoaderHelper<K, V> {
 
     DispatchResult<V> dispatch() {
         boolean batchingEnabled = loaderOptions.batchingEnabled();
-        //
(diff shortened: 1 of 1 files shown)


bbakerman (approved):
Thanks for this PR - this helps reduce memory pressure

PR #147, opened 2024-04-30
Verify a throwing CacheMap#get does not break DataLoader

AlexandreCarlton:
As a fast follow to #146, we add a regression test to ensure that we still handle a `CacheMap` whose `#get` throws an exception (by falling back to the underlying batch loader).


bbakerman (approved):
Great idea

PR #155, opened 2024-05-21
Migrate publisher tests

AlexandreCarlton:
In this pull request we:

 - move across the `*BatchPublisher` tests into `DataLoaderTest` to obtain the same coverage with less code.
 - fix `MappedBatchPublisher` loader to function without a cache (which would otherwise not work as we were completing only the last future found for a given key).


bbakerman (approved):
Nice one

PR #160, opened 2024-05-26
Have MappedBatchPublisher take in a Set<K> keys (and add README sections)

AlexandreCarlton:
This is more symmetric with `MappedBatchLoader` and preserves efficiency; we do not need to emit a `Map.Entry` for duplicate keys (given the strong intention that this will be used to create a `Map`).

This was picked up after creating `README` examples, which have also been added here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,66 @@ For example, let's assume you want to load users from a database, you could prob
         // ...
 ```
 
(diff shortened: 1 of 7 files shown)


bbakerman (approved):
Nice one

PR #171, opened 2025-01-01
Parameterise more tests across DataLoader implementations

AlexandreCarlton:
We parameterise:

 - `DataLoaderValueCacheTest`
 - `ScheduledDataLoaderRegistryTest`

to provide increased coverage across all DataLoader implementations.

Notably, `DataLoaderValueCacheTest` is _not_ parameterised across the Publisher DataLoaders (yet) as these do not appear to work when a `ValueCache` is provided (which is what prompted this pull request). It will be fixed in a future pull request, at which point we will restore coverage.

### DIFF
--- a/src/test/java/org/dataloader/DataLoaderValueCacheTest.java
+++ b/src/test/java/org/dataloader/DataLoaderValueCacheTest.java
@@ -4,10 +4,13 @@ import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import org.dataloader.fixtures.CaffeineValueCache;
 import org.dataloader.fixtures.CustomValueCache;
+import org.dataloader.fixtures.parameterized.TestDataLoaderFactory;
 import org.dataloader.impl.DataLoaderAssertionException;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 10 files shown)


src/test/java/org/dataloader/fixtures/parameterized/TestDataLoaderFactories.java
```diff
@@ -0,0 +1,25 @@
+package org.dataloader.fixtures.parameterized;
+
+import org.junit.jupiter.api.Named;
+import org.junit.jupiter.params.provider.Arguments;
+
+import java.util.stream.Stream;
+
+public class TestDataLoaderFactories {
+    public static Stream<Arguments> get() {
```
AlexandreCarlton, line 9: Lifted this to a central location so it could be used across multiple tests.

src/test/java/org/dataloader/DataLoaderValueCacheTest.java
```diff
@@ -30,11 +32,12 @@
 
 public class DataLoaderValueCacheTest {
 
-    @Test
-    public void test_by_default_we_have_no_value_caching() {
-        List<List<String>> loadCalls = new ArrayList<>();
+    @ParameterizedTest
+    @MethodSource("org.dataloader.fixtures.parameterized.TestDataLoaderFactories#getWithoutPublisher")
```
AlexandreCarlton, line 36: As mentioned in the description, we use `getWithoutPublisher` rather than `get` as `ValueCache` does not appear to interact well with the `Publisher` `DataLoader`s.

src/test/java/org/dataloader/fixtures/TestKit.java
```diff
@@ -61,43 +60,14 @@ public static <K, V> BatchLoader<K, V> keysAsValues(List<List<K>> loadCalls) {
         };
     }
 
-    public static <K, V> BatchLoader<K, V> keysAsValuesAsync(Duration delay) {
```
AlexandreCarlton, line 64: Figured I'd clean up the now unused methods - we should be using the `TestDataLoaderFactory`-provided `DataLoader`s anyway, if we feel the need to reach out for these older ones.

src/test/java/org/dataloader/fixtures/parameterized/TestDataLoaderFactory.java
```diff
@@ -19,4 +22,17 @@ public interface TestDataLoaderFactory {
     DataLoader<String, String> onlyReturnsNValues(int N, DataLoaderOptions options, ArrayList<Object> loadCalls);
 
     DataLoader<String, String> idLoaderReturnsTooMany(int howManyMore, DataLoaderOptions options, ArrayList<Object> loadCalls);
+
+    // Convenience methods
```
AlexandreCarlton, line 26: Really, this was just to minimise the diff.

bbakerman (approved):

PR #252, opened 2025-11-22
Add strict mode for DataLoaderRegistry

AlexandreCarlton:
In a similar vein to https://github.com/graphql-java/graphql-java/pull/3565 which enabled "strict mode" for type wiring (preventing multiple `DataFetcher`s being registered to the same field on a GraphQL type), we add "strict mode" to the `DataLoaderRegistry` so that we don't accidentally register multiple DataLoaders to the same key (which would result in the last registration winning). This can prevent confusing bugs from emerging.

This defaults to `false` to avoid breaking changes, and deliberately mimics the referenced PR to maintain consistency.

### DIFF
--- a/src/main/java/org/dataloader/DataLoaderRegistry.java
+++ b/src/main/java/org/dataloader/DataLoaderRegistry.java
@@ -85,7 +85,10 @@ public class DataLoaderRegistry {
      * @param existingDL              the existing data loader
      * @return a new {@link DataLoader} or the same one if there is nothing to change
      */
-    private static DataLoader<?, ?> nameAndInstrumentDL(String key, @Nullable DataLoaderInstrumentation registryInstrumentation, DataLoader<?, ?> existingDL) {
+    private DataLoader<?, ?> nameAndInstrumentDL(String key, @Nullable DataLoaderInstrumentation registryInstrumentation, DataLoader<?, ?> existingDL) {
+        if (strictMode) {
+            assertKeyStrictly(key);
+        }
         existingDL = checkAndSetName(key, existingDL);
 
         if (registryInstrumentation == null) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/dataloader/DataLoaderRegistry.java
```diff
@@ -144,6 +153,9 @@ private static DataLoaderOptions setInInstrumentation(DataLoaderOptions options,
      */
     public DataLoaderRegistry register(DataLoader<?, ?> dataLoader) {
         String name = Assertions.nonNull(dataLoader.getName(), () -> "The DataLoader must have a non null name");
+        if (strictMode) {
+            assertKeyStrictly(name);
+        }
         dataLoaders.put(name, nameAndInstrumentDL(name, instrumentation, dataLoader));
```
bbakerman, line 159: You need to cover `org.dataloader.DataLoaderRegistry#computeIfAbsent` as well

I think this code would be better put in the `nameAndInstrumentDL` method as a common place

The following is all the places it is called



```
instrumentDLs(Map<String, DataLoader<?, ?>>, DataLoaderInstrumentation)
register(DataLoader<?, ?>)
register(String, DataLoader<?, ?>)
registerAndGet(String, DataLoader<?, ?>)
computeIfAbsent(String, Function<String, DataLoader<?, ?>>)
```

bbakerman (changes requested):
Move the strict check in `nameAndInstrumentDL` as a common place and all the `registering` methods will then be covered

src/main/java/org/dataloader/DataLoaderRegistry.java
AlexandreCarlton: Done - as agreed elsewhere, I've omitted `computeIfAbsent` since this repeated invocations to the same `key` will not override the `DataLoader` with strict mode off.

I _did_ have to make `nameAndInstrumentDL` to get access to `strictMode` and the `DataLoader` map, however.

bbakerman (approved):
Thanks

PR #261, opened 2026-01-06
Add dispatch strategy, implement bfs strategy

zsmoore:
See discussion in - https://github.com/graphql-java/java-dataloader/discussions/256
Build on https://github.com/graphql-java/java-dataloader/pull/243

Allow for custom dispatch strategies

Right now the dispatch strategy (in this PR and the original busy strategy) do not need dataloader references, they just need registry references.  Its possible that strategies prefer to have the DL loaded such as a DFS strategy to trigger the dl dispatch immediately without using a registry reference but as there is no current need do not include it.

Because of the circular reference dependency between dispatch strategy and registry, choose to add a bootstrap function into the interface to give a reference to the strategy

BFS strategy handles dispatching in a level by level fashion see example here - 

i.e.

Async Path

    A
        B (async)
            E
            F
        C
            G (async)
            H
        D
            I
            J

A dispatches immediately, queues B, C, D
B, C, D dispatch queue G, H, I, J
Thread is spawned to wait for B to finish and retry every 30 ms
B finishes, queues E, F
E, F, G, H, I, J dispatched

This dispatch strategy also handles chained dataloaders but beyond that handles ASYNC chained dataloaders such that a dataloader makes an api call then chains a subsequent dataloader.

This is done with minimal thread overload by doing dispatchAll + only queueing a single thread when work is known to exist but not be completed

Added some tests to verify some behavior of the dispatch strategy


zsmoore:
Note to self to follow up on - I believe thinking on this again this may be a greedy level by level traversal rather than true BFS.  I will come back to this and add a more complicated test to replicate the example I have to verify if async are waited on a level basis or we continue dispatching to child levels when work is known and only wait if we are blocked on just async calls

src/main/java/org/dataloader/strategy/GreedyLevelByLevelChainedDispatchStrategy.java
```diff
@@ -0,0 +1,131 @@
...
+                    scheduleFallbackDispatch();
+                    break;
+                }
+
+                // completed
+                if (pendingAfter == 0) {
+                    resetState();
+                }
+            }
+        }
```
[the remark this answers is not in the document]
zsmoore: I don't agree with this review comment as its an ineffective suggestion.

If we throw in the dispatch strategy the exception is not propagated upward as expected as we cannot intercept the future that the dataloader user is expecting to complete.  Beyond this if the loop iteration is stuck in the fallback thread this complicates the issue even more.

This leaved the user a slightly ambiguous experience as their future can hang if they introduce an infinite loop into the tree but i think this is acceptable and we cannot provide a better experience when handling general dispatch.

zsmoore:
> Note to self to follow up on - I believe thinking on this again this may be a greedy level by level traversal rather than true BFS. I will come back to this and add a more complicated test to replicate the example I have to verify if async are waited on a level basis or we continue dispatching to child levels when work is known and only wait if we are blocked on just async calls

Wrote out a fairly complicated test in `GreedyLevelByLevelChainedDispatchStrategyStressTest` to verify the behavior in practice.  As a result renamed this as its not a true BFS.  In practice we see increased batch sizes compared to our previous ticker mode with scheduled registry so while this is not a true level by level traversal we still see a great balance of speed + efficiency in our real life scenario

zsmoore:
@andimarek This one should be about ready.  Let me know what you think when you're available to review

src/main/java/org/dataloader/strategy/GreedyLevelByLevelChainedDispatchStrategy.java
```diff
@@ -0,0 +1,162 @@
...
+ * chained dataloaders. This minimizes the amount of threads spawned to only be used when there is no known work to be
+ * done and the chain is not finished.
+ * <p>
+ * Due to the concept of 'known' work we greedily walk the chain instead of waiting for async calls to finish before
+ * kicking off the next level.
+ * <p>
+ * In practice this will greedily fill up DataLoader keys while walking the chain to provide a nice balance of
+ * batching/dedupe/caching while not needing to worry about manually dispatching the tree.
+ */
+public class GreedyLevelByLevelChainedDispatchStrategy implements DispatchStrategy {
```
bbakerman, line 34: This class needs the following

```
@PublicApi
@NullMarked
```

src/main/java/org/dataloader/DataLoaderHelper.java
```diff
@@ -188,14 +188,18 @@ CompletableFuture<V> load(K key, Object loadContext) {
         }
 
         ctx.onDispatched();
+        loaderOptions.getDispatchStrategy().loadCalled();
         loadCallFuture.whenComplete(ctx::onCompleted);
+        loadCallFuture.whenComplete((result, error) -> loaderOptions.getDispatchStrategy().loadCompleted());
```
bbakerman: This will create 2 idependent `whenCompletes`

That might be what we want but I thought I woud ask

if we want `ctx::onCompleted` to happen always before before `loaderOptions.getDispatchStrategy().loadCompleted()` then we need to have one `whenComplete` and make it dependent.

I suspect we should only have one.  With `loaderOptions.getDispatchStrategy().loadCompleted()` done first and then `ctx::onCompleted` say

src/main/java/org/dataloader/DataLoaderHelper.java
```diff
@@ -188,14 +188,18 @@ CompletableFuture<V> load(K key, Object loadContext) {
...
         return loadCallFuture;
     }
 
     private CompletableFuture<V> incrementCacheHitAndReturnCF(DataLoaderInstrumentationContext<Object> ctx, K key, Object loadContext, CompletableFuture<V> cachedFuture) {
         stats.incrementCacheHitCount(new IncrementCacheHitCountStatisticsContext<>(key, loadContext));
         ctx.onDispatched();
+        loaderOptions.getDispatchStrategy().loadCalled();
         cachedFuture.whenComplete(ctx::onCompleted);
+        cachedFuture.whenComplete((result, error) -> loaderOptions.getDispatchStrategy().loadCompleted());
```
bbakerman: Same here - should they be independent or dependent?

bbakerman (approved):
I am happy with this PR - modulo the question on whenComplete ordering

bbakerman:
ps the test are failing

```
GreedyLevelByLevelChainedDispatchStrategyTest > singleDepthLoadSucceedsMultipleTimes() FAILED
    java.lang.NullPointerException: Cannot invoke "java.lang.Runnable.run()" because "this.onIteration" is null
        at org.dataloader.strategy.GreedyLevelByLevelChainedDispatchStrategy.triggerDeterministicDispatch(GreedyLevelByLevelChainedDispatchStrategy.java:82)
        at org.dataloader.strategy.GreedyLevelByLevelChainedDispatchStrategy.loadCalled(GreedyLevelByLevelChainedDispatchStrategy.java:61)
        at org.dataloader.DataLoaderHelper.load(DataLoaderHelper.java:191)
        at org.dataloader.DataLoader.loadImpl(DataLoader.java:202)
        at org.dataloader.DataLoader.load(DataLoader.java:216)
        at 
```

andimarek:
> ps the test are failing
> 
> ```
> GreedyLevelByLevelChainedDispatchStrategyTest > singleDepthLoadSucceedsMultipleTimes() FAILED
>     java.lang.NullPointerException: Cannot invoke "java.lang.Runnable.run()" because "this.onIteration" is null
>         at org.dataloader.strategy.GreedyLevelByLevelChainedDispatchStrategy.triggerDeterministicDispatch(GreedyLevelByLevelChainedDispatchStrategy.java:82)
>         at org.dataloader.strategy.GreedyLevelByLevelChainedDispatchStrategy.loadCalled(GreedyLevelByLevelChainedDispatchStrategy.java:61)
>         at org.dataloader.DataLoaderHelper.load(DataLoaderHelper.java:191)
>         at org.dataloader.DataLoader.loadImpl(DataLoader.java:202)
>         at org.dataloader.DataLoader.load(DataLoader.java:216)
>         at 
> ```

I fixed that.

andimarek:
@zsmoore @bbakerman I changed the strategy callback a bit to ensure it is always called before the load itself is returned. 

@zsmoore the stress test is failing now (not sure if that was the result of my change, it could be).

I will continue to review.

andimarek:
@zsmoore some questions, clarifications: 

In your test your "chained dataloader" is actually a few DataLoader.load calls in parallel. What we refer normally to as chained dataloader is 
```java
dl1.load(key1).compose(result1 -> var key2 <based on result1)>; return dl2.load(key2))
```

In `CF` terms we compose the result of 1 DL to call a second one.

We also have a `loadMany` function that would server the as the same purpose I believe.

In your test `dispatchGoesByLevel` you do actually something we have not really seen before I believe: you use a new DataLoader inside the BatchLoader itself, correct? You do it twice inside "root" and inside "secondLevel".

This is again a bit different than the use cased described above we called "chained".

So that is what you are referring to as "level"?

Regarding your strategy:

In my understanding it works like this:
First load triggers dispatch. The dispatch keep calling dispatch as long as there is work todo. You keep a lock for the whole dispatch time. In case the work is not done synchronously it will triggers a delayed dispatch after 30ms to dispatch all DataLoader that were asynchronously queued.

zsmoore:
@bbakerman @andimarek apologies for the delay, I was ill the past couple of weeks.

Changes since last commits - 
1 - Folded the 2 whenCompletes into 1, i agree we can keep the order deterministic for the callbacks this way
2 - Changes to tests WRT equality, dependencies, and exception unfolding  Why?  With the latest change which made the returned future the chained '.whenComplete' future, we now cannot use an equality assertion as the future is not pass through.  This is most likely the preferred behavior as we won't lose exceptions and will bubble them up.

zsmoore:
> @zsmoore some questions, clarifications:
> 
> In your test your "chained dataloader" is actually a few DataLoader.load calls in parallel. What we refer normally to as chained dataloader is
> 
> ```java
> dl1.load(key1).compose(result1 -> var key2 <based on result1)>; return dl2.load(key2))
> ```
> 
> In `CF` terms we compose the result of 1 DL to call a second one.
> 
> We also have a `loadMany` function that would server the as the same purpose I believe.
> 
> In your test `dispatchGoesByLevel` you do actually something we have not really seen before I believe: you use a new DataLoader inside the BatchLoader itself, correct? You do it twice inside "root" and inside "secondLevel".
> 
> This is again a bit different than the use cased described above we called "chained".
> 
> So that is what you are referring to as "level"?
> 
> Regarding your strategy:
> 
> In my understanding it works like this: First load triggers dispatch. The dispatch keep calling dispatch as long as there is work todo. You keep a lock for the whole dispatch time. In case the work is not done synchronously it will triggers a delayed dispatch after 30ms to dispatch all DataLoader that were asynchronously queued.

For the understanding of the strategy yes that is the correct understanding of how the strategy is working.


As for the verbage around chained I see the point around the mixing of chaining.  We actually have both cases in our use case take for example

Nested chaining
First Name DataLoader (input profile nickname) ->
  Call Profile DataLoader (input profile nickname)
  return profile.firstName
  
Top level chaining
(Get Location)
Call Profile DataLoader.compose(call location dataloader using profile id)

When referring to level, I think of the entire call chain as a tree structure.

Take both example above

Nested Chaining
Root function = 0
Call FirstName DL = 1
Call Profile DL = 2

Top level Chaining
Root function = 0
Call ProfileDL = 1
Call LocationDL = 2

In both cases the 'level' or depth of the chain is the same as it takes the same amount of subsequent DL dependents to get the overall result but i believe the difference comes down to organization.  You could still swap the structure to be 
LocationDL (input profile nickname)
  Call ProfileDL (find by profile nickname)
  use profileID to call Location endpoint and return
  
Where this composition is encapsulated in the overall DL itself instead of a top level unfolding of the calls.

Let me know if i understood the question correctly.  Thanks again and apologies for the delay

zsmoore:
Also WRT the parallelization, that is just added to bump up the complexity for the stress test, nothing really core to the test here

zsmoore:
@bbakerman @andimarek friendly ping on this one

andimarek:
hey @zsmoore ... sorry it has been really busy the last week. Will try to come back to this soon.

zsmoore:
No worries at all, the beginning of the year is always crazy

dondonz:
Reopening - I wanted to chat about this before the bot closes it

kaqqao:
This bot is a scourge...

PR #275, opened 2026-05-17
Allow scheduling of CF completion - Attempt 1

bbakerman:
The completion of CF promises can now be scheduled on another thread

A step in the batch loading process is the completion of the futures that have been promised earlier via
`Dataloader.load()`. This can also be scheduled on another thread if need be, allowing the dispatch call 
to return quicker.

You might want to schedule the completion of values on another thread if there is following work t do such as :

```java
var cfStage1 = dataLoader.load(key);
var cfStage2 = cfStage1.thenApply(v -> doSomethingSlow(v));
//...
var dispatchCF = dataLoader.dispatch();
```     

In the above example the `.doSomethingSlow(v)` call will happen inside the completion
of the `cfStage1` code path since `CompletableFuture` by design eagerly runs dependent chained methods.

Perhaps you want this completion to happen more asynchronously so that the `.dispatch()` returns more
quickly and is not bound to the `.doSomethingSlow(v)` call.

By default, the dispatch completion is done on the current thread in a synchronous manner, which will include
any extra `CompletableFuture` chained methods.

This is an example of running the completion step in an asynchronous manner :

```java

@Override
public <K> CompletionStage<?> scheduleCompletion(Runnable completeValuesRunnable, List<K> keys, BatchLoaderEnvironment environment) {
    return CompletableFuture.runAsync(completeValuesRunnable);
}
```

### DIFF
--- a/src/main/java/org/dataloader/DataLoaderHelper.java
+++ b/src/main/java/org/dataloader/DataLoaderHelper.java
@@ -242,7 +242,7 @@ class DataLoaderHelper<K, V> {
         int queueSize = loaderQueueEntryHead.queueSize;
         // we copy the pre-loaded set of futures ready for dispatch
         Object[] keysArray = new Object[queueSize];
-        CompletableFuture[] queuedFuturesArray = new CompletableFuture[queueSize];
+        CompletableFuture<V>[] queuedFuturesArray = new CompletableFuture[queueSize];
         Object[] callContextsArray = new Object[queueSize];
         int index = queueSize - 1;
         while (loaderQueueEntryHead != null) {
--- a/src/main/java/org/dataloader/DataLoaderHelper.java
+++ b/src/main/java/org/dataloader/DataLoaderHelper.java
@@ -19,7 +19,6 @@ import java.time.Clock;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
@@ -32,7 +31,6 @@ import java.util.concurrent.atomic.AtomicReference;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
-import static java.util.concurrent.CompletableFuture.allOf;
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static java.util.stream.Collectors.toList;
 import static org.dataloader.impl.Assertions.assertState;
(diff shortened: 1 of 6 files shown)


src/main/java/org/dataloader/DataLoaderHelper.java
```diff
@@ -313,54 +311,56 @@ private CompletableFuture<List<V>> sliceIntoBatchesOfBatches(List<K> keys, List<
...
     }
 
     @SuppressWarnings("unchecked")
-    private CompletableFuture<List<V>> dispatchQueueBatch(List<K> keys, List<Object> callContexts, List<CompletableFuture<V>> queuedFutures) {
-        stats.incrementBatchLoadCountBy(keys.size(), new IncrementBatchLoadCountByStatisticsContext<>(keys, callContexts));
-        CompletableFuture<List<V>> batchLoad = invokeLoader(keys, callContexts, queuedFutures, loaderOptions.cachingEnabled());
+    private CompletableFuture<List<V>> dispatchQueueBatch(List<K> keys, List<Object> keyContexts, List<CompletableFuture<V>> queuedFutures) {
+        stats.incrementBatchLoadCountBy(keys.size(), new IncrementBatchLoadCountByStatisticsContext<>(keys, keyContexts));
+
+        BatchLoaderEnvironment environment = mkBatchLoaderEnv(keys, keyContexts);
```
bbakerman, line 321: We now create the BatchLoaderEnvironment earlier and send it down into the other methods earlier

src/main/java/org/dataloader/DataLoaderHelper.java
```diff
@@ -313,54 +311,56 @@ private CompletableFuture<List<V>> sliceIntoBatchesOfBatches(List<K> keys, List<
...
-                            stats.incrementLoadErrorCount(new IncrementLoadErrorCountStatisticsContext<>(key, callContext));
-                            future.completeExceptionally((Throwable) value);
-                            clearCacheKeys.add(keys.get(idx));
-                        } else if (value instanceof Try) {
-                            // we allow the batch loader to return a Try so we can better represent a computation
-                            // that might have worked or not.
-                            Try<V> tryValue = (Try<V>) value;
-                            if (tryValue.isSuccess()) {
-                                future.complete(tryValue.get());
-                            } else {
```
bbakerman, line 351: The above is now run in the runnable and hence can be scheduled onto some other thread

src/main/java/org/dataloader/DataLoaderHelper.java
```diff
@@ -313,54 +311,56 @@ private CompletableFuture<List<V>> sliceIntoBatchesOfBatches(List<K> keys, List<
...
+                        List<K> clearCacheKeys = new ArrayList<>();
+                        for (int idx = 0; idx < queuedFutures.size(); idx++) {
+                            K key = keys.get(idx);
+                            V value = values.get(idx);
+                            Object callContext = keyContexts.get(idx);
+                            CompletableFuture<V> future = queuedFutures.get(idx);
+                            if (value instanceof Throwable) {
                                 stats.incrementLoadErrorCount(new IncrementLoadErrorCountStatisticsContext<>(key, callContext));
-                                future.completeExceptionally(tryValue.getThrowable());
+                                future.completeExceptionally((Throwable) value);
```
bbakerman, line 341: diff is whacky.  This has not changed

src/main/java/org/dataloader/DataLoaderHelper.java
```diff
@@ -375,6 +375,26 @@ private CompletableFuture<List<V>> dispatchQueueBatch(List<K> keys, List<Object>
...
+        return scheduledCompletion
+                .thenApply(ignored -> values)
+                .toCompletableFuture();
+    }
+
+    private BatchLoaderEnvironment mkBatchLoaderEnv(List<K> keys, List<Object> keyContexts) {
+        Object context = loaderOptions.getBatchLoaderContextProvider().getContext();
+        return BatchLoaderEnvironment.newBatchLoaderEnvironment()
+                .context(context).keyContexts(keys, keyContexts).build();
+    }
```
bbakerman, line 396: Called in two places - common method

src/test/java/org/dataloader/impl/CompletableFutureKitTest.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class CompletableFutureKitTest {
```
bbakerman, line 15: never had tests

src/main/java/org/dataloader/impl/CompletableFutureKit.java
```diff
@@ -22,6 +23,12 @@ public static <V> CompletableFuture<V> failedFuture(Exception e) {
         return future;
     }
 
+    public static <V> CompletableFuture<V> success(V v) {
```
fxbonnet, line 26: Isn't this the same as `CompletableFuture.completedFuture(v)`?

src/main/java/org/dataloader/DataLoaderHelper.java
fxbonnet: This moved the CF completions to another thread but the for loop still invokes them sequentially which is a problem if there are many keys in the batch. Could we schedule the completion for each key in a separate thread instead?

src/main/java/org/dataloader/impl/CompletableFutureKit.java
bbakerman: Yes - but here for "completeness" ;)

src/main/java/org/dataloader/DataLoaderHelper.java
bbakerman: So are you saying we pass in

```
 public CompleableFuture<Void> completeThese(List<CompletableFuture> cfs, List<Object> values)
 ```
 
 And the the caller could sequentially run them as today OR choose to run them in parralel themselves say?

PR #277, opened 2026-05-18
Allow scheduling of CF completion in parallel

fxbonnet:
When a batch loading function returns, we currently complete all `CompletableFuture`s sequentially on the calling thread. Since `CompletableFuture.complete(...)` also executes any continuations attached via `thenApply(...)` or `thenCompose(...)`, large batches can significantly delay the return of `dispatch()`.

This PR introduces an option to execute these continuations in parallel instead.

### DIFF
--- a/src/main/java/org/dataloader/DataLoaderHelper.java
+++ b/src/main/java/org/dataloader/DataLoaderHelper.java
@@ -19,6 +19,7 @@ import java.time.Clock;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
@@ -27,6 +28,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicReference;
 
 import static java.util.Collections.emptyList;
@@ -329,17 +331,18 @@ class DataLoaderHelper<K, V> {
                         return CompletableFutureKit.success(values);
                     }
 
-                    Runnable completeValuesRunnable = () -> {
-                        List<K> clearCacheKeys = new ArrayList<>();
-                        for (int idx = 0; idx < queuedFutures.size(); idx++) {
-                            K key = keys.get(idx);
-                            V value = values.get(idx);
-                            Object callContext = keyContexts.get(idx);
-                            CompletableFuture<V> future = queuedFutures.get(idx);
+                    Collection<K> clearCacheKeys = new ConcurrentLinkedQueue<>();
+                    List<Runnable> completeValueRunnables = new ArrayList<>();
+                    for (int idx = 0; idx < queuedFutures.size(); idx++) {
+                        K key = keys.get(idx);
+                        V value = values.get(idx);
+                        Object callContext = keyContexts.get(idx);
+                        CompletableFuture<V> future = queuedFutures.get(idx);
+                        Runnable completeValueRunnable = () -> {
                             if (value instanceof Throwable) {
                                 stats.incrementLoadErrorCount(new IncrementLoadErrorCountStatisticsContext<>(key, callContext));
                                 future.completeExceptionally((Throwable) value);
-                                clearCacheKeys.add(keys.get(idx));
+                                clearCacheKeys.add(key);
                             } else if (value instanceof Try) {
                                 // we allow the batch loader to return a Try so we can better represent a computation
                                 // that might have worked or not.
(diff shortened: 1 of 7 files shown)


src/main/java/org/dataloader/DataLoaderHelper.java
```diff
@@ -325,40 +326,54 @@ private CompletableFuture<List<V>> dispatchQueueBatch(List<K> keys, List<Object>
...
+                                Try<V> tryValue = (Try<V>) value;
+                                if (tryValue.isSuccess()) {
+                                    future.complete(tryValue.get());
+                                } else {
+                                    stats.incrementLoadErrorCount(new IncrementLoadErrorCountStatisticsContext<>(key, callContext));
+                                    future.completeExceptionally(tryValue.getThrowable());
+                                    clearCacheKeys.add(key);
+                                }
+                            } else {
+                                future.complete(value);
```
bbakerman, line 358: This misses the `Try` path of completion

andimarek:
@fxbonnet I am not sure about this PR: with the new hook introduced by @bbakerman in #275 you can already offload every completion onto a new thread and hence it will run in parallel if you choose todo so. I have problems seeing how this approach is different/better.

fxbonnet:
@andimarek I created this PR after my comment on @bbakerman 's PR https://github.com/graphql-java/java-dataloader/pull/275#discussion_r3255674227
The PR allowed to offload the completions to a separate thread, but in case the batch load contains multiple keys, the completions for each key would still happen sequentially.
For clarity I just rebased this PR onto @bbakerman 's branch. The main change is that instead of passing a single `Runnable` to `scheduleCompletion`, we would pass a `List<Runnable>` (one for each key), giving the option to execute them in parallel.

src/main/java/org/dataloader/DataLoaderHelper.java
```diff
@@ -313,54 +313,59 @@ private CompletableFuture<List<V>> sliceIntoBatchesOfBatches(List<K> keys, List<
...
+                .thenCompose(values -> {
                     assertResultSize(keys, values);
                     if (isPublisher() || isMappedPublisher()) {
                         // We have already completed the queued futures by the time the overall batchLoad future has completed.
-                        return values;
+                        return CompletableFutureKit.success(values);
                     }
 
-                    List<K> clearCacheKeys = new ArrayList<>();
+                    Collection<K> clearCacheKeys = new ConcurrentLinkedQueue<>();
```
fxbonnet, line 334: `clearCacheKeys` may now be accessed concurrently by multiple thread if we execute the `Runnable`s in parallel.

bbakerman: good call - price of side effects outside a thread

bbakerman (approved):

andimarek (changes requested):
Requesting changes for a semantic regression around `ValueCache` hits and context-aware batch loaders.

I opened #278 as a small draft PR showing one possible fix: https://github.com/graphql-java/java-dataloader/pull/278

The PR now constructs a `BatchLoaderEnvironment` before value-cache filtering:

```java
BatchLoaderEnvironment environment = mkBatchLoaderEnv(keys, keyContexts);
CompletableFuture<List<V>> batchLoad = invokeLoader(environment, keys, keyContexts, queuedFutures, loaderOptions.cachingEnabled());
```

But `invokeLoader(..., cachingEnabled=true)` can then remove value-cache hits and invoke the actual batch loader with only `missedKeys` / `missedKeyContexts` while still passing the original environment:

```java
CompletableFuture<List<V>> batchLoad = invokeLoader(environment, missedKeys, missedKeyContexts, missedQueuedFutures);
```

That breaks the existing semantics of `BatchLoaderEnvironment`: `environment.getKeyContextsList()` is expected to be aligned with the `keys` argument passed to the loader. A concrete failure case:

1. The queued dispatch contains `keys=[A, B]` and `keyContexts=[ctxA, ctxB]`.
2. `ValueCache` returns a hit for `A` and a miss for `B`.
3. The actual batch loader is invoked with `keys=[B]`.
4. The reused environment was built from `[A, B]`, so `environment.getKeyContextsList().get(0)` is `ctxA`, not `ctxB`.

That means a `BatchLoaderWithContext` implementation that indexes `environment.getKeyContextsList()` alongside its `keys` argument can load `B` using `A`'s context. This is observable behavior and can be a security/data-isolation issue if key contexts carry tenant, auth, locale, or request-scoped metadata.

Please build the `BatchLoaderEnvironment` from the actual key/context list passed to the concrete loader call after value-cache filtering, or otherwise preserve the invariant that `environment.getKeyContextsList()` is ordered and sized to match the loader's `keys` argument. This should also get a regression test with a partial `ValueCache` hit and a context-aware batch loader asserting that the missed key receives its own context.

PR #278 does exactly that by rebuilding the environment for `missedKeys` / `missedKeyContexts` before the concrete loader invocation and adding the regression test.

andimarek:
I opened a small draft PR with one possible fix for the ValueCache / BatchLoaderEnvironment context alignment issue: #278

fxbonnet:
Thanks @andimarek I have added your commit to the PR

bbakerman:
This should now be ready - it has the extra fix @andimarek wanted

fxbonnet:
Hi @andimarek is there something I need to do for this PR to be ready to merge?

Issue #273, opened 2026-04-02
Issue in reactive data-loader


zenios:
https://github.com/graphql-java/java-dataloader/blob/e5e5bf58ef2d3eeebaa9b994bd43e0cc7527ed8f/src/main/java/org/dataloader/reactive/AbstractBatchSubscriber.java#L53

When the subscription happens we request the size of the keys.

Lets take an example where we batch load one key. 
The implementation is buggy and the result is two rows. But the second is the one matching the actual key requesting.

Once the first result is received, the observable will wait for a signal to send more which never happens and the data loader is staying blocked.

The solution is to keep keep track of pending requests and when it reaches zero we should re-request until we eventually get to onError or onComplete.

At the same time if we manage to receive all the values that we expect prior to completing upstream we are free to cancel upstream and complete manually

bbakerman:
Sorry missed this issue.  We would need a test showing the bug happening - I cant work out the problem from the description above.

PR #279, opened 2026-06-16
Fix reactive data loader hanging on backpressured publishers (#273)

zenios:
The reactive subscribers (`AbstractBatchSubscriber` and friends) requested `keys.size()` from the upstream publisher exactly once in `onSubscribe` and then waited for `onComplete` to finalise the batch.

A reactive publisher only emits while it has outstanding demand. So a publisher that emits values lazily as more demand arrives - for example one that emits an entry not matching any requested key before the entry that does - consumes that single window of demand and then blocks. The matching value is never requested, the per-key future never completes, and the data loader hangs forever.

This change makes the subscribers manage demand properly:

  * track the outstanding demand and re-request another window whenever it drains to zero, repeating until the publisher completes or errors;
  * once a result has been received for every key there is nothing left to wait for, so cancel the upstream subscription and complete ourselves rather than blocking on a publisher that may never call `onComplete`.

Adds `ReactiveBackpressureTest` reproducing the hang for both the list and mapped publisher data loaders; both tests time out against the old code and pass with the fix.

### DIFF
--- a/src/main/java/org/dataloader/reactive/AbstractBatchSubscriber.java
+++ b/src/main/java/org/dataloader/reactive/AbstractBatchSubscriber.java
@@ -34,6 +34,11 @@ abstract class AbstractBatchSubscriber<K, V, T> implements Subscriber<T> {
     boolean onErrorCalled = false;
     boolean onCompleteCalled = false;
 
+    // the upstream subscription and how much demand we currently have outstanding with it.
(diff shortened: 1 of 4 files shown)


bbakerman (approved):
Thanks very much for this.   Demand management is an art and often gotten wrong.

I also like the idea we can ignore keys we didn't ask for.  Probably not likely but possible and hence handling it is a good idea

